PR #120, opened 2021-05-27
fix: await parachain confirmations in redeem/replace/refund

sander2:


nud3l (approved):

vault/src/execution.rs
```diff
@@ -164,7 +175,16 @@ impl Request {
...
+        tracing::info!("Awaiting parachain confirmations...");
+
+        provider
+            .wait_for_block_in_relay(
+                H256Le::from_bytes_le(&tx_metadata.block_hash.to_vec()),
+                Some(num_confirmations),
+            )
+            .await?;
+
+        tracing::info!("Bitcoin successfully sent and relayed");
```
nud3l, line 187: minor nit: the recipient is quite interesting in the logs, but not blocking for this PR

gregdhill (approved):

PR #190, opened 2021-10-28
feat(vault): wrapped currency CLA

savudani8:
- adds a CLA for specifying the wrapped currency of the vault
- fixes a bug where the Bitcoin RPC passed for listening to issue requests is walletless. Vaults register a new Bitcoin deposit address on each issue request, and since the Bitcoin RPC connection was walletless this operation failed and caused the client to crash. The fix is to dynamically determine the vault-specific Bitcoin RPC on each `request_issue` event, based on the `vault_id`


vault/src/system.rs
```diff
@@ -98,7 +98,11 @@ pub struct VaultServiceConfig {
 
     /// The currency to use for the collateral, e.g. "DOT" or "KSM".
     #[clap(long, parse(try_from_str = parse_collateral_currency))]
-    pub currency_id: Option<CurrencyId>,
+    pub collateral_currency_id: Option<CurrencyId>,
```
sander2, line 101: I think there is some clap option to require another arg if you pass this one, something like `#[clap(long, requires("wrapped-currency-id"), parse(try_from_str = parse_collateral_currency))]`, or possibly `requires = "wrapped-currency-id"`. Could you try that?

vault/src/system.rs
```diff
@@ -548,13 +552,16 @@ impl VaultService {
...
-                tracing::info!("Not registering vault -- currency-id not configured");
-                return Ok(());
-            }
-        };
+        let (collateral_currency, _wrapped_currency) =
+            match (self.config.collateral_currency_id, self.config.wrapped_currency_id) {
+                (Some(x), Some(y)) => (x, y),
+                _ => {
+                    tracing::info!(
+                    "Not registering vault -- collateral-currency-id and wrapped-currency-id must both be configured"
```
sander2, line 560: with the change requested above, this would become `"Not registering vault -- collateral-currency-id and wrapped-currency-id not configured"`. Note that we don't require the currency to be set currently - if none is specified, we just don't register. If the user then registers the vault through the ui, the client will pick it up

vault/src/system.rs
savudani8: Used `require`, but it seems a bit buggy. It logs the same missing argument three times now:
```
error: The following required arguments were not provided:
    --collateral-currency-id <COLLATERAL_CURRENCY_ID>
    --collateral-currency-id <COLLATERAL_CURRENCY_ID>
    --collateral-currency-id <COLLATERAL_CURRENCY_ID>
```

vault/src/issue.rs
```diff
@@ -193,25 +193,29 @@ async fn add_new_deposit_key<B: BitcoinCoreApi + Clone + Send + Sync + 'static>(
...
     let event_channel = &event_channel;
     let issue_set = &issue_set;
+    let btc_rpc = &btc_rpc;
     btc_parachain
         .on_event::<RequestIssueEvent, _, _, _>(
             |event| async move {
                 if &event.vault_id.account_id == btc_parachain.get_account_id() {
+                    let bitcoin_core = match btc_rpc.get_bitcoin_rpc(&event.vault_id).await {
+                        Some(x) => x,
+                        None => return, // event not directed at this vault
```
sander2, line 211: This is a good fix, but in the future please either create a separate PR for unrelated fixes, or if there is a reason not to, then at least mention it in the PR description so we don't miss it in the review. Anyway, let's print an error here - we already checked above that the event is directed at this accountid - if we can't find the bitcoinrpc, it indicates a bug.

vault/src/system.rs
sander2: Hmm weird. Probably fine for now though

sander2 (approved):

PR #195, opened 2021-12-02
fix: prevent refund double spending

sander2:


nud3l (commented):
Is it actually possible to add an integration test for this? Following case I have in mind:

- Request issue and overpay BTC
- After the request issue, Vault is exactly at the secure thresholds (can e.g., just adjust the exchange rate to bring down the vault to the desired collateralization level)
- Let the first refund try fail with insufficient collateral
- Increase the collateral of the Vault to have enough collateral to cover the refund_fee
- Expect the refund request to succeed with no extra BTC transaction

On a related note, I'm wondering if it actually made sense to issue interBTC from the refund_fe eor if the vault should just keep the BTC.

sander2:
> Is it actually possible to add an integration test for this? Following case I have in mind:
> 
>     * Request issue and overpay BTC
> 
>     * After the request issue, Vault is exactly at the secure thresholds (can e.g., just adjust the exchange rate to bring down the vault to the desired collateralization level)
> 
>     * Let the first refund try fail with insufficient collateral
> 
>     * Increase the collateral of the Vault to have enough collateral to cover the refund_fee
> 
>     * Expect the refund request to succeed with no extra BTC transaction

This is a test we can add, but it does not actually test the fix that I made here - this only modifies the startup behavior.

>  On a related note, I'm wondering if it actually made sense to issue interBTC from the refund_fee or if the vault should just keep the BTC.

The problem is that they can't spend it, it'd marked as theft

nud3l (approved):

PR #197, opened 2022-01-04
fix: update parachain version (with subxt fix)

sander2:
- new metadata
- use the new currencyid type
- the parachain upgrade exposed a new subxt bug; upgrade to a new version that includes [the fix I made there](https://github.com/paritytech/subxt/pull/382)


gregdhill (approved):

gregdhill (commented):
Instead of conditionally compiling the faucet we could only compile it with standalone metadata, but with the new testnet runtime sudo will be available again. See: https://github.com/interlay/interbtc/pull/465

PR #203, opened 2022-01-17
fix: faucet currency

sander2:
This will make the faucet send KINT in addition to whatever currency is requested


vault/src/faucet.rs
```diff
@@ -21,7 +21,8 @@ where
 
 #[derive(Encode, Decode, Debug, Clone, serde::Serialize)]
 struct FundAccountJsonRpcRequest {
-    pub vault_id: VaultId,
+    pub vault_account_id: AccountId,
+    pub collateral_currency: CurrencyId,
```
sander2, line 25: I noticed this struct didn't match the definition inside the faucet. It would be nice if we could actually test making the http request in an integration test..

PR #232, opened 2022-02-17
feat: vault MVP metrics

savudani8:
- Adds prometheus crate and exposes a `/metrics` endpoint where metrics can be fetched by prometheus
- Includes configuration for setting up a local prometheus service that pings the vault
- Metrics include:
  - Bitcoin balance (updated on each in / out transaction)
  - Average BTC fee (fetched on service startup and updated using the `TransactionMetadata`)
  - collateralization
  - locked collateral
  - required collateral (to resume to the secure threshold)
  - process_cpu_seconds_total
  - process_max_fds
  - process_open_fds
  - process_resident_memory_bytes
  - process_virtual_memory_bytes
- bridge-specific metrics get updated on each `FeedValues` oracle event (every 25 mins). These are quite costly as they call parachain RPCs
- includes a minimal Grafana template for the metrics

Adds three new flags, identical to the ones used for monitoring Substrate nodes:
```bash
        --no-prometheus
            Do not expose a Prometheus metric endpoint
        --prometheus-external
            Expose Prometheus exporter on all interfaces.
            
            Default is local.

        --prometheus-port <PROMETHEUS_PORT>
            Specify Prometheus exporter TCP Port
            
            [default: 9615]
```
---
# Steps to run Prometheus and Grafana
Prometheus:
```bash
cd vault/src/monitoring
docker build -t prometheus .
docker run -p 9090:9090 --network=host prometheus
```

Grafana installation and running instructions (works by default with the Prometheus setup): https://grafana.com/docs/grafana/latest/installation/debian/#2-start-the-server
# Example Visualisation
Screenshot from 2022-03-14 17-55-22


vault/src/main.rs
```diff
@@ -35,18 +35,24 @@ async fn start() -> Result<(), Error> {
...
         opts.parachain,
         opts.service,
         opts.vault,
-    )
-    .start()
-    .await?;
+    );
+
+    let (_, future_result) = futures::join!(
+        warp::serve(metrics_route).run(([0, 0, 0, 0], 8080)),
```
gregdhill, line 51: Let's make this configurable, use a similar CLI flag to whatever Substrate uses.

vault/src/main.rs
```diff
@@ -35,18 +35,24 @@ async fn start() -> Result<(), Error> {
...
+        tokio::task::spawn(async move { vault_connection_manager.start().await }),
+    );
 
-    Ok(())
+    future_result?
```
gregdhill, line 55: Do you not need to `.await` this?

runtime/src/rpc.rs
```diff
@@ -1611,6 +1613,23 @@ impl VaultRegistryPallet for InterBtcParachain {
 
         Ok(result.amount)
     }
+
+    async fn get_collateralization_from_vault(&self, vault_id: VaultId, only_issued: bool) -> Result<u128, Error> {
```
gregdhill, line 1617: Do we still use this RPC in the lib? Wondering if we should just do the calculation here since we anyway read the storage values.

savudani8: The logic is duplicated in the lib because of issues with RPC calls in polkadot-js. We should revert to using the RPC call once fixed, so best to keep it in the parachain.

vault/src/main.rs
savudani8: The `join` above implicitly awaits (similar to `(a.await, b.await)`)

savudani8:
Grafana dashboard. The collateral currency token can be read from the data point legend (too verbose to fit in the screenshot). 

Screenshot from 2022-03-04 17-15-28

vault/src/metrics.rs
```diff
@@ -0,0 +1,172 @@
...
+    if bitcoin_fees_vec.len() == 0 && new_fee_entry.is_none() {
+        // Initialize the `AVERAGE_BTC_FEE` by fetching all Bitcoin txs.
+        match btc_rpc.list_transactions(None).await {
+            Ok(tx_list) => {
+                let mut fees: Vec<_> = tx_list
+                    .iter()
+                    .filter_map(|tx| tx.detail.fee.map(|amount| amount.as_btc()))
+                    .collect();
+
+                bitcoin_fees_vec.append(&mut fees);
```
sander2, line 124: Unless you plan to use this vec for other statistics, there is no need no maintain this vec - just maintain two variables containing the total amount, and the number of transactions.

vault/src/metrics.rs
```diff
@@ -0,0 +1,172 @@
...
+                for currency_id in updated_currencies {
+                    match vault_ids
+                        .iter()
+                        .find(|vault_id| &vault_id.collateral_currency() == currency_id)
+                    {
+                        None => tracing::debug!("Ignoring exchange rate update for {}", currency_id.inner().symbol()),
+                        Some(vault_id) => {
+                            tracing::info!("Received FeedValuesEvent for {}", currency_id.inner().symbol());
+                            if let Err(err) = update_bridge_metrics(parachain_rpc.clone(), vault_id.clone()).await {
+                                tracing::info!("{:?}", err);
```
sander2, line 162:
```suggestion
                                tracing::info!("Failed to update bridge metrics: {}", err);
```

vault/src/metrics.rs
```diff
@@ -0,0 +1,172 @@
...
+                            }
+                        }
+                    }
+                }
+            },
+            |error| tracing::error!("Error reading SetExchangeRate event: {}", error.to_string()),
```
sander2, line 168:
```suggestion
            |error| tracing::error!("Error reading FeedValues event: {}", error.to_string()),
```

vault/src/system.rs
```diff
@@ -649,13 +675,25 @@ impl VaultService {
             .bitcoin_theft_start_height
             .unwrap_or(self.bitcoin_core.get_block_count().await? as u32 + 1);
 
-        let vaults_listener = wait_or_shutdown(
+        let vault_id = self.get_vault_id();
+        let bitcoin_core_with_wallet = match self.vault_id_manager.get_bitcoin_rpc(&vault_id).await {
+            Some(btc_rpc) => btc_rpc,
+            None => self.bitcoin_core.clone(),
```
sander2, line 681: We should abort here

vault/src/system.rs
```diff
@@ -649,13 +675,25 @@ impl VaultService {
...
+            self.btc_parachain.get_bitcoin_confirmations().await?,
+        )
+        .await;
+
+        let bitcoin_listener = wait_or_shutdown(
             self.shutdown.clone(),
-            report_vault_thefts(
-                self.bitcoin_core.clone(),
+            monitor_btc_txs(
+                bitcoin_core_with_wallet.clone(),
```
sander2, line 692: I think this breaks multi collateral support

vault/src/metrics.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+pub async fn update_bitcoin_metrics<B: BitcoinCoreApi + Clone + Send + Sync + 'static>(
+    btc_rpc: B,
+    btc_confirmations: u32,
+) {
+    match btc_rpc.get_balance(Some(btc_confirmations)).await {
+        Ok(bitcoin_balance) => LOCKED_BTC.set(bitcoin_balance.as_btc()),
+        Err(e) => {
+            // failed to cancel; get up-to-date request list in next iteration
```
sander2, line 103: comment doesn't make sense

vault/src/vaults.rs
```diff
@@ -136,9 +156,10 @@ impl<P: RelayPallet + BtcRelayPallet, B: BitcoinCoreApi + Clone> VaultTheftMonit
             bitcoin::stream_in_chain_transactions(self.bitcoin_core.clone(), self.btc_height, num_confirmations).await;
 
         while let Some(Ok((block_hash, tx))) = stream.next().await {
-            if let Err(err) = self.check_transaction(tx, block_hash, num_confirmations).await {
+            if let Err(err) = self.check_transaction(tx.clone(), block_hash, num_confirmations).await {
                 tracing::error!("Failed to check transaction: {}", err);
             }
+            self.monitor_bitcoin_metrics(tx, num_confirmations).await;
```
sander2, line 162: This doesn't make a lot of sense to me - why do we listen on the bitcoin chain for our own transactions? Why not just call `update_bitcoin_metrics` directly after making a transaction?

sander2 (changes requested):

sander2 (approved):

vault/src/vaults.rs
savudani8: I'm thinking that not all bitcoin transactions may be made through this client. What if the client crashes and a redeem payment is made manually?

PR #252, opened 2022-03-14
refactor: optimize vault theft scanning

sander2:
The theft reporting was very slow due to inefficient rpc usage. In particular, for every transaction in the bitcoin chain, we made 3 rpc calls: `parachain.wait_for_block_in_relay`, `bitcoin.get_raw_tx` and `bitcoin.get_proof`.

With this change, we only call `wait_for_block_in_relay` once per block rather than once per transaction. Furthermore, we now call `get_raw_tx` only for transactions that are done by vaults, and only call `get_proof` when the transaction is found to be invalid.


vault/src/vaults.rs
```diff
@@ -69,26 +76,42 @@ impl<P: RelayPallet + BtcRelayPallet, B: BitcoinCoreApi + Clone> VaultTheftMonit
...
                 if !self.btc_parachain.is_transaction_invalid(vault_id, &raw_tx_2).await? {
-                    tracing::info!("Found a double payment - reporting...");
+                    tracing::info!(
+                        "Detected double payment by vault {} - txids {} and {}. Reporting...",
+                        vault_id.pretty_printed(),
+                        transaction.txid(),
+                        txid_2
+                    );
+                    let proof_1 = self.bitcoin_core.get_proof(transaction.txid(), &block_hash).await?;
+                    let proof_2 = self.bitcoin_core.get_proof(txid_2, &block_hash_2).await?;
```
gregdhill, line 110: Probably doesn't make sense to parallelize since underlying implementation is synchronous but maybe we can leave a comment here?

vault/src/vaults.rs
```diff
@@ -69,26 +76,42 @@ impl<P: RelayPallet + BtcRelayPallet, B: BitcoinCoreApi + Clone> VaultTheftMonit
...
-            self.btc_parachain.report_vault_theft(vault_id, proof, raw_tx).await?;
+        let raw_tx = bitcoin::serialize(transaction); //self.bitcoin_core.get_raw_tx(&tx_id, &block_hash).await?;
+        if self.btc_parachain.is_transaction_invalid(vault_id, &raw_tx).await? {
+            tracing::info!(
+                "Detected theft by vault {} - txid {}. Reporting...",
+                vault_id.pretty_printed(),
+                transaction.txid()
+            );
+
+            let proof = self.bitcoin_core.get_proof(transaction.txid(), &block_hash).await?;
```
gregdhill, line 96: Should we retry? Not expecting changes now, but maybe leave a TODO note here.

vault/src/vaults.rs
sander2: > Probably doesn't make sense to parallelize since underlying implementation is synchronous but maybe we can leave a comment here?

I don't think there's a need - if there are so many double payments that this becomes a bottleneck, then we have a bigger problem on our hands

vault/src/vaults.rs
```diff
@@ -99,44 +122,69 @@ impl<P: RelayPallet + BtcRelayPallet, B: BitcoinCoreApi + Clone> VaultTheftMonit
...
+                .then(move |block| async move {
+                    let transactions: Box<dyn Stream<Item = _> + Unpin + Send> = match block {
+                        Ok(e) => {
+                            let block_hash = e.block_hash();
+                            // at this point we know that the transaction has `num_confirmations` on the bitcoin chain,
+                            // but the relay can introduce a delay, so wait until the relay also confirms the
+                            // transaction.
+                            let transactions: Box<dyn Stream<Item = _> + Unpin + Send> = match self
+                                .btc_parachain
+                                .wait_for_block_in_relay(
```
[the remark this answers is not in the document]
sander2: Yea I guess it might be possible. I think we need to review the entire code for correct fork handling, I think I saw some todos that mentioned that forks might not be handled correctly. Let's make a separate ticket for that?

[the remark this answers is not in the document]
gregdhill: Yeah can you make a ticket for this?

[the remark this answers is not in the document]
sander2: [ticket](https://www.notion.so/interlay/Check-correct-bitcoin-fork-handling-in-vault-client-d854a04ddd25454ab4aa165e16aca732)

vault/src/vaults.rs
sander2: Done, left a comment on the (renamed) `check_vault_transaction` function

gregdhill (approved):

PR #254, opened 2022-03-17
refactor: monitor per currencypair & misc refactorings

sander2:
This PR is on top of https://github.com/interlay/interbtc-clients/pull/232. Feel free to change the base of this PR to master if you want an overview of the total changes compared to master.

- Each vaultid now uses its own copy of the per-currencypair metrics. Usage is now e.g. `locked_btc{currency="KSM_KBTC", instance="localhost:9615", job="service-collector"}` 
- Instead of scanning the bitcoin chain for our own payments, now update metrics directly after we make payments
- Optimize average bitcoin fee tracking


vault/src/metrics.rs
```diff
@@ -0,0 +1,518 @@
...
+
+        for vault in vault_id_manager.get_entries().await {
+            publish_utxo_count(&vault).await;
+        }
+
+        sleep(SLEEP_DURATION).await;
+    }
+}
+
+pub async fn update_expected_bitcoin_balance<B: BitcoinCoreApi + Clone + Send + Sync>(
```
sander2, line 508: should change this to btc instead of satoshi

bitcoin/src/lib.rs
```diff
@@ -455,6 +462,23 @@ impl BitcoinCoreApi for BitcoinCore {
...
+    async fn get_balance(&self, min_confirmations: Option<u32>) -> Result<Amount, Error> {
+        Ok(self
+            .rpc
+            .get_balance(min_confirmations.map(|x| x.try_into().unwrap_or_default()), None)?)
+    }
+
+    /// List the transaction in the wallet.
+    async fn list_transactions(&self, max_count: Option<usize>) -> Result<Vec<json::ListTransactionResult>, Error> {
+        // Arbitrarily picked number for the max tx count to receive. The bitcoin rpc default is 10.
+        let default_max_count = 100_000_000;
```
gregdhill, line 475: This should probably be a constant.

bitcoin/src/lib.rs
```diff
@@ -807,6 +841,10 @@ impl BitcoinCoreApi for BitcoinCore {
             .collect();
         Ok(ret?)
     }
+
+    async fn get_utxo_count(&self) -> Result<usize, Error> {
```
gregdhill, line 845: Please add doc comment.

vault/src/metrics.rs
```diff
@@ -1,42 +1,249 @@
...
+
+const CURRENCY_LABEL: &str = "currency";
+const BTC_BALANCE_TYPE_LABEL: &str = "type";
+const REQUEST_STATUS_LABEL: &str = "status";
 
-// Metrics are stored under the `vault_id` key so that multiple vaults can be easily
+// Metrics are stored under the [`CURRENCY_LABEL`] key so that multiple vaults can be easily
 // monitored at the same time.
 lazy_static! {
-    static ref BITCOIN_FEES: RwLock<Vec<f64>> = RwLock::new(vec![]);
```
savudani8: My reasoning for storing an array was that we might want to track outliers or other statistics, so I was wary of over-optimizing

vault/src/system.rs
```diff
@@ -183,9 +209,21 @@ impl<BCA: BitcoinCoreApi + Clone + Send + Sync + 'static> VaultIdManager<BCA> {
...
 
-        self.bitcoin_rpcs.write().await.insert(vault_id, btc_rpc.clone());
+        tracing::info!("Initializing metrics...");
+        let metrics = PerCurrencyMetrics::new(&vault_id);
+        let data = VaultData {
+            vault_id: vault_id.clone(),
+            btc_rpc: btc_rpc.clone(),
+            metrics: metrics.clone(),
+        };
+        tracing::info!("Initializing metrics..222.");
```
savudani8, line 223: typo?
```suggestion
        tracing::info!("Initializing metrics...");
```

runtime/src/integration/bitcoin_simulator.rs
```diff
@@ -240,19 +241,32 @@ impl BitcoinCoreApi for MockBitcoinCore {
             sleep(Duration::from_secs(1)).await;
         }
     }
+    async fn get_balance(&self, min_confirmations: Option<u32>) -> Result<Amount, BitcoinError> {
+        Ok(Amount::ZERO)
```
savudani8, line 245: Why not keep the `unimplemented!()`?

sander2: I think the tests were failing because of it

vault/src/metrics.rs
```diff
@@ -0,0 +1,527 @@
...
+        Err(e) => {
+            // unexpected error, but not critical so just continue
+            tracing::warn!("Failed to get Bitcoin balance: {}", e);
+        }
+    }
+}
+
+async fn publish_kint_balance(parachain_rpc: &InterBtcParachain) {
+    if let Ok(balance) = parachain_rpc.get_free_balance(parachain_rpc.native_currency_id).await {
+        KINT_BALANCE.set(balance as f64 / parachain_rpc.native_currency_id.one() as f64);
```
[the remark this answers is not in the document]
sander2: Really? Wouldn't that make things more difficult? Since e.g. intr and kint don't have the same number of decimals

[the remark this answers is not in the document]
savudani8: In this case the BTC balance and fees should also be in Satoshi. Grafana seems to allow setting custom Y-axis units https://grafana.com/docs/grafana/latest/visualizations/time-series/change-axis-display/

vault/src/metrics.rs
```diff
@@ -0,0 +1,527 @@
...
+        publish_kint_balance(&parachain_rpc).await;
+        publish_issue_count(&parachain_rpc.clone(), &vault_id_manager).await;
+        publish_redeem_count(&parachain_rpc.clone(), &vault_id_manager).await;
+
+        for vault in vault_id_manager.get_entries().await {
+            publish_utxo_count(&vault).await;
+        }
+
+        sleep(SLEEP_DURATION).await;
+    }
```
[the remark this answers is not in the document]
sander2: True. The reason I initially went for this approach is because I wanted to have 2 different counters for expired & cancelled, and we don't have parachain events for redeem expiration. In the end I forgot to actually implement it this way though. Anyway, I would propose to leave it like this for now and leave it as a future optimization

vault/src/metrics.rs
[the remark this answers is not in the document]
sander2: Because it shifts the burden of making the conversions to the vault operator - they will need to know and modify the number of decimals for each currency they use. I don't really understand why you want to take away the automation and rely on the operators instead - what is the advantage?

vault/src/metrics.rs
```diff
@@ -0,0 +1,527 @@
...
+                .await;
+
+            *vault.metrics.fee_budget_surplus.data.write().await = fee_budget_surplus;
+            publish_fee_budget_surplus(&vault).await;
+        }
+
+        publish_utxo_count(&vault).await;
+        publish_average_bitcoin_fee(&vault).await;
+        publish_bitcoin_balance(&vault).await;
+        update_expected_bitcoin_balance(&vault, parachain_rpc.clone()).await;
```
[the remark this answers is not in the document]
sander2: Not really worth it - as you noted, two of these are not actually async due to bitcoin rpc being sync, and one of the others is only async because of the rwlock

vault/src/metrics.rs
```diff
@@ -0,0 +1,527 @@
...
+}
+
+async fn publish_kint_balance(parachain_rpc: &InterBtcParachain) {
+    if let Ok(balance) = parachain_rpc.get_free_balance(parachain_rpc.native_currency_id).await {
+        KINT_BALANCE.set(balance as f64 / parachain_rpc.native_currency_id.one() as f64);
+    }
+}
+
+async fn publish_utxo_count<B: BitcoinCoreApi + Clone + Send + Sync>(vault: &VaultData<B>) {
+    if let Ok(count) = vault.btc_rpc.get_utxo_count().await {
```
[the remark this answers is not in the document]
sander2: Good point, will update the function to non-async

PR #259, opened 2022-03-23
fix: add timeout on transaction submission

sander2:
This adds a 5 minute timeout on all transaction submissions. Since I am not sure of subxt/jsonrpsee deals well with cancelled subscriptions, and because a timeout might indicate an invalid state in those dependencies, I made _all_ timeouts fire the shutdown signal so a fresh connection will get established upon restart. Without this explicit shutdown signal, non-essential calls (such as bitcoin block relay) would not cause a restart.

The reason for the long timeout time is so that it's less likely the stuck transaction will get included into the parachain after we timeout.

### DIFF
--- a/faucet/src/main.rs
+++ b/faucet/src/main.rs
@@ -64,7 +64,9 @@ async fn main() -> Result<(), Error> {
     let faucet_config = opts.faucet;
 
     loop {
-        let btc_parachain = parachain_config.try_connect(signer.clone()).await?;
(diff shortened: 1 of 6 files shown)


nud3l:
Caused by upstream issues

jsonrpsee: https://github.com/paritytech/jsonrpsee/issues/714
substrate: https://github.com/paritytech/substrate/issues/10910

gregdhill (approved):

PR #270, opened 2022-03-31
fix: issue cancellation: take into account issuerequest's period

sander2:
This fixes the bug where, after the global issue period just having decreased, the vault tried to cancel old issues prematurely. The fix is to take into account the issue request's period in addition to the global period.

### DIFF
--- a/vault/src/cancellation.rs
+++ b/vault/src/cancellation.rs
@@ -24,7 +24,6 @@ pub enum Event {
 pub struct CancellationScheduler<P: IssuePallet + ReplacePallet + UtilFuncs + Clone> {
     parachain_rpc: P,
     vault_id: AccountId,
-    period: Option<u32>,
     parachain_height: BlockNumber,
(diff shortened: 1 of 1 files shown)


vault/src/cancellation.rs
```diff
@@ -307,19 +311,6 @@ impl<P: IssuePallet + ReplacePallet + UtilFuncs + SecurityPallet + Clone> Cancel
 
         Ok(ret)
     }
-
-    /// Cached function to get the issue/replace period, in number of blocks until
-    /// it is allowed to be canceled
-    async fn get_cached_period<T: Canceller<P>>(&mut self) -> Result<u32, Error> {
```
sander2, line 313: We need to refetch in case the period change, so we can't cache forever. The performance difference is anyway negligible because the period is only fetched after having fetched all open issues, which is more expensive.

gregdhill (approved):

PR #307, opened 2022-06-21
feat: allow oracle to use a combination of fixed and real exchange rates

sander2:
Will be required on interlay. Before, we required either all or none exchange rates to be fixed. Now it is possible to mix both options.

Usage is e.g. `--coingecko=https://api.coingecko.com/api/v3/ --currency-id INTR=6649 --currency-id DOT`


gregdhill (approved):
Nice API, very clean.

PR #311, opened 2022-07-20
feat(pruned-client): implementation of pruned bitcoin node support in the vault, with electrs querying for issue requests

theodorebugnet:
As per title.

Outstanding question is integration testing. Bitcoin refuses to prune blocks for a very long time - even when issuing a manual `pruneblockchain` rpc call. It doesn't seem realistically possible to create a situation with a pruned transaction during a test run - not without either forking bitcoin-core to see if this behaviour can be patched for more aggressive pruning, or investigating if manually deleting block data may work. A "normal" integration test for the vault would just mock bitcoin-core to return "success" which seems totally useless given that the critical functionality is specifically importing the tx into the bitcoin wallet, hence this PR currently has no tests. I'm open to discussion on this.

### DIFF
--- a/bitcoin/src/electrs.rs
+++ b/bitcoin/src/electrs.rs
@@ -1,4 +1,7 @@
 use crate::Error;
+use esplora_btc_api::models::Transaction;
+
+const ELECTRS_TRANSACTIONS_PER_PAGE: usize = 25;
 
 async fn get(base_path: &str, url: &str) -> Result<String, Error> {
     let response = reqwest::get(format!("{}/{}", base_path, url)).await?;
@@ -10,10 +13,29 @@ async fn get(base_path: &str, url: &str) -> Result<String, Error> {
     }
 }
 
+// the esplora_btc_api lib breaks on these two because it tries to serde_json::from_str on a plain string value
 pub async fn get_tx_hex(base_path: &str, tx_id: &str) -> Result<String, Error> {
     get(base_path, &format!("tx/{}/hex", tx_id)).await
 }
 
 pub async fn get_tx_merkle_block_proof(base_path: &str, tx_id: &str) -> Result<String, Error> {
     get(base_path, &format!("tx/{}/merkleblock-proof", tx_id)).await
+}
+
+// and it doesn't currently support the paged api call
+pub async fn get_address_tx_history_full(base_path: &str, address: &str) -> Result<Vec<Transaction>, Error> {
+    let mut last_seen_txid = "".to_owned();
+    let mut ret = Vec::<Transaction>::default();
+    loop {
+        let mut transactions: Vec<Transaction> =
+            serde_json::from_str(&get(base_path, &format!("address/{}/txs/chain/{}", address, last_seen_txid)).await?)?;
+        let page_size = transactions.len();
+        last_seen_txid = transactions.last().map_or("", |tx| &tx.txid).to_owned();
+        ret.append(&mut transactions);
+        if page_size < ELECTRS_TRANSACTIONS_PER_PAGE {
+            // no further pages
+            break;
+        }
+    }
+    Ok(ret)
 }
--- a/bitcoin/src/electrs.rs
+++ b/bitcoin/src/electrs.rs
@@ -1,0 +1,19 @@
+use crate::Error;
+
+async fn get(base_path: &str, url: &str) -> Result<String, Error> {
+    let response = reqwest::get(format!("{}/{}", base_path, url)).await?;
+    let status = response.status();
+    if status.is_client_error() || status.is_server_error() {
+        Err(Error::ElectrsQueryFailed)
+    } else {
+        Ok(response.text().await?)
+    }
+}
+
+pub async fn get_tx_hex(base_path: &str, tx_id: &str) -> Result<String, Error> {
+    get(base_path, &format!("tx/{}/hex", tx_id)).await
+}
+
+pub async fn get_tx_merkle_block_proof(base_path: &str, tx_id: &str) -> Result<String, Error> {
+    get(base_path, &format!("tx/{}/merkleblock-proof", tx_id)).await
+}
--- a/bitcoin/src/iter.rs
+++ b/bitcoin/src/iter.rs
@@ -20,10 +20,9 @@ use std::iter;
 pub async fn reverse_stream_transactions<B: BitcoinCoreApi + Clone + Send + Sync + 'static>(
     rpc: &B,
     stop_height: u32,
-    stop_at_pruned: bool,
 ) -> Result<impl Stream<Item = Result<Transaction, Error>> + Unpin + '_, Error> {
     let mempool_transactions = stream::iter(rpc.get_mempool_transactions().await?);
-    let in_chain_transactions = reverse_stream_in_chain_transactions(rpc, stop_height, stop_at_pruned).await;
+    let in_chain_transactions = reverse_stream_in_chain_transactions(rpc, stop_height).await;
     Ok(mempool_transactions.chain(in_chain_transactions))
 }
 
@@ -38,19 +37,16 @@ pub async fn reverse_stream_transactions<B: BitcoinCoreApi + Clone + Send + Sync
 pub async fn reverse_stream_in_chain_transactions<B: BitcoinCoreApi + Clone + Send + Sync + 'static>(
     rpc: &B,
     stop_height: u32,
-    stop_at_pruned: bool,
 ) -> impl Stream<Item = Result<Transaction, Error>> + Send + Unpin + '_ {
-    reverse_stream_blocks(rpc, stop_height, stop_at_pruned)
-        .await
-        .flat_map(|block| {
-            // unfortunately two different iterators don't have compatible types, so we have
-            // to box them to trait objects
-            let transactions: Box<dyn Stream<Item = _> + Unpin + Send> = match block {
-                Ok(e) => Box::new(stream::iter(e.txdata.into_iter().map(Ok))),
-                Err(e) => Box::new(stream::iter(iter::once(Err(e)))),
-            };
-            transactions
-        })
+    reverse_stream_blocks(rpc, stop_height).await.flat_map(|block| {
+        // unfortunately two different iterators don't have compatible types, so we have
+        // to box them to trait objects
+        let transactions: Box<dyn Stream<Item = _> + Unpin + Send> = match block {
+            Ok(e) => Box::new(stream::iter(e.txdata.into_iter().map(Ok))),
+            Err(e) => Box::new(stream::iter(iter::once(Err(e)))),
+        };
+        transactions
+    })
 }
 
 /// Stream blocks in reverse order, starting at the current best height reported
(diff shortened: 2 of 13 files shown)


bitcoin/src/lib.rs
```diff
@@ -880,6 +893,46 @@ impl BitcoinCoreApi for BitcoinCore {
...
     }
 
+    async fn rescan_electrs_for_addresses<A: PartialAddress + Send + Sync + 'static>(
+        &self,
+        addresses: Vec<A>,
+    ) -> Result<(), Error> {
+        for address in addresses.into_iter() {
+            let address = address.encode_str(self.network)?;
+            let all_transactions =
+                esplora_btc_api::apis::address_api::get_address_tx_history(&self.electrs_config, &address).await?;
```
gregdhill, line 903: Can you add a TODO note that this should be paged? I think this call will fail if there are too many addresses.

bitcoin/src/lib.rs
```diff
@@ -880,6 +893,46 @@ impl BitcoinCoreApi for BitcoinCore {
...
+            let confirmed_payments_to = all_transactions.into_iter().filter(|tx| {
```
gregdhill, line 904: Leave a comment specifying what this array contains

bitcoin/src/lib.rs
```diff
@@ -880,6 +893,46 @@ impl BitcoinCoreApi for BitcoinCore {
...
+                if let Some(status) = &tx.status {
```
gregdhill, line 905: When is this value `None`? Should we return false if doesn't exist?

vault/src/issue.rs
```diff
@@ -92,23 +92,54 @@ pub async fn add_keys_from_past_issue_request<B: BitcoinCoreApi + Clone + Send +
         None => return Ok(()), // the iterator is empty so we have nothing to do
     };
 
-    for (issue_id, request) in issue_requests.into_iter() {
+    for (issue_id, request) in issue_requests.clone().into_iter() {
         if let Err(e) = add_new_deposit_key(bitcoin_core, issue_id, request.btc_public_key).await {
             tracing::error!("Failed to add deposit key #{}: {}", issue_id, e.to_string());
         }
+        // TODO: get TXs from elects
+        // then importprunedfunds
```
gregdhill, line 100: This seems like something we want in this PR right?

vault/src/issue.rs
```diff
@@ -92,23 +92,54 @@ pub async fn add_keys_from_past_issue_request<B: BitcoinCoreApi + Clone + Send +
...
+    tracing::info!("Rescanning bitcoin chain from height {}...", rescan_start_height);
+    for (range_start, range_end) in chunks(rescan_start_height, btc_end_height) {
         tracing::debug!("Scanning chain blocks {range_start}-{range_end}...");
         bitcoin_core.rescan_blockchain(range_start, range_end).await?;
     }
 
+    // also check in electrs in case there were any requests from before the pruned height
+    if btc_start_height < btc_pruned_start_height {
+        tracing::info!(
+            "Also checking electrs for isssue requests between {} and {}...",
```
gregdhill, line 123:
```suggestion
            "Also checking electrs for issue requests between {} and {}...",
```

gregdhill (commented):
Looking good, only major point is whether we implement the same import logic when subscribed to issue events.

bitcoin/src/lib.rs
theodorebugnet: Actually, you're right (assuming you meant too many transactions). I operated under the assumption that each issue address will only ever be used once, but that does open up an attack where anyone can spam 25 dust-amount transactions to an issue request, and then this API call will no longer pick up the original payment.

Only an issue if the vault client isn't running in the first place while the real issue transaction occurs, but definitely should be mitigated, I'll work on adding payment. Good catch, thanks.

vault/src/issue.rs
theodorebugnet: Whoops, old dev comment. (We do want that - and it's already implemented of course, haha.)

bitcoin/src/lib.rs
theodorebugnet: Apparently the `Transaction` type from electrs has most of its fields as an `Options`, not sure why.  I don't see in what case it would be None, but in any case we only import confirmed transactions and those would have `status: { confirmed: true }`. The API seems to always return the status object just fine.

theodorebugnet:
> Looking good, only major point is whether we implement the same import logic when subscribed to issue events.

What do you mean - as in switching to looking up electrs for new issue requests that come in _while_ the client is running? If so then I don't really see the sense in that, we're always gonna have the very recent transactions available in bitcoin core and it's more reliable and faster to query the local daemon than try to make a remote network request.

sander2:
Have you looked into the pruneblockchain rpc for testing?

bitcoin/src/lib.rs
```diff
@@ -880,6 +893,49 @@ impl BitcoinCoreApi for BitcoinCore {
...
+                if let Some(vout) = &tx.vout {
+                    for output in vout {
+                        if let Some(addr) = &output.scriptpubkey_address {
+                            if addr == &address {
+                                return true;
+                            }
+                        }
+                    }
+                };
+                false
```
sander2, line 922:
```suggestion
        // check that the address is used as _output_, since presumably bitcoin core already knows about transactions only _spend from_ the address 
        tx.vout
            .unwrap_or_default()
            .iter()
            .any(|output| matches!(output.scriptpubkey_address, Some(addr) if addr == address));
```

bitcoin/src/electrs.rs
```diff
@@ -0,0 +1,19 @@
...
+    let response = reqwest::get(format!("{}/{}", base_path, url)).await?;
+    let status = response.status();
+    if status.is_client_error() || status.is_server_error() {
+        Err(Error::ElectrsQueryFailed)
+    } else {
+        Ok(response.text().await?)
+    }
+}
+
+pub async fn get_tx_hex(base_path: &str, tx_id: &str) -> Result<String, Error> {
```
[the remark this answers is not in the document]
gregdhill: Unfortunately the code generated by the OpenAPI spec always uses `serde_json::from_str` regardless of the actual return type and I couldn't see a way to change this behaviour. This fails to decode a `String` since it is not valid JSON and there is no way to patch the library right now since the code is generated in the CI.

bitcoin/src/iter.rs
```diff
@@ -30,19 +33,24 @@ pub async fn reverse_stream_transactions<B: BitcoinCoreApi + Clone + Send + Sync
 ///
 /// * `rpc` - bitcoin rpc
 /// * `stop_height` - height of the last block the iterator will return transactions from
+/// * `stop_at_pruned` - whether to gracefully stop if a pruned blockchain is encountered;
+/// otherwise, will throw an error
 pub async fn reverse_stream_in_chain_transactions<B: BitcoinCoreApi + Clone + Send + Sync + 'static>(
     rpc: &B,
     stop_height: u32,
+    stop_at_pruned: bool,
```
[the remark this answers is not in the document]
theodorebugnet: Good point, my thinking when putting this in was that we might want to disable pruned support (e.g. if someone really doesn't want external electrs calls) and thus force fail on encountering a pruned node, but that doesn't seem likely and if it ever becomes necessary we can fail early elsewhere. Removing.

vault/src/issue.rs
```diff
@@ -103,12 +103,41 @@ pub async fn add_keys_from_past_issue_request<B: BitcoinCoreApi + Clone + Send +
...
-    tracing::info!("Rescanning bitcoin chain from height {}...", btc_start_height);
-    for (range_start, range_end) in chunks(btc_start_height, btc_end_height) {
+    // check if the blockchain was pruned after the point where we would need to scan
+    // if it was, we only rescan from the pruned height
+    let btc_pruned_start_height = bitcoin_core.get_pruned_height().await? as usize;
+    let rescan_start_height = btc_start_height.max(btc_pruned_start_height);
+
+    // in parallel, rescan what blockchain we do have stored locally
+    tracing::info!("Rescanning bitcoin chain from height {}...", rescan_start_height);
+    for (range_start, range_end) in chunks(rescan_start_height, btc_end_height) {
```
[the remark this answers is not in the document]
theodorebugnet: I thought it best to minimise the amount of network calls and maximise local usage. If we remove the scanning altogether, this will affect _all_ vaults and make them query electrs for all their issue requests - is that something we want to do?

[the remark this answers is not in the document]
gregdhill: I think it should still be possible for the Vault to be operated *without* electrs to reduce our dependency on that service.

bitcoin/src/electrs.rs
```diff
@@ -10,10 +11,29 @@ async fn get(base_path: &str, url: &str) -> Result<String, Error> {
...
+pub async fn get_address_tx_history_full(base_path: &str, address: &str) -> Result<Vec<Transaction>, Error> {
+    let mut last_seen_txid = "".to_owned();
+    let mut ret = Vec::<Transaction>::default();
+    loop {
+        let mut transactions: Vec<Transaction> =
+            serde_json::from_str(&get(base_path, &format!("address/{}/txs/chain/{}", address, last_seen_txid)).await?)?;
+        let page_size = transactions.len();
+        last_seen_txid = transactions.last().map_or("", |tx| &tx.txid).to_owned();
+        ret.append(&mut transactions);
+        if page_size < 25 {
```
sander2, line 33: Let's make this a named constant?

bitcoin/src/electrs.rs
```diff
@@ -10,10 +11,29 @@ async fn get(base_path: &str, url: &str) -> Result<String, Error> {
...
     get(base_path, &format!("tx/{}/merkleblock-proof", tx_id)).await
 }
+
+// and it doesn't currently support the paged api call
+pub async fn get_address_tx_history_full(base_path: &str, address: &str) -> Result<Vec<Transaction>, Error> {
+    let mut last_seen_txid = "".to_owned();
+    let mut ret = Vec::<Transaction>::default();
+    loop {
+        let mut transactions: Vec<Transaction> =
+            serde_json::from_str(&get(base_path, &format!("address/{}/txs/chain/{}", address, last_seen_txid)).await?)?;
```
sander2, line 29: Nit: shouldn't the trailing `/` be omitted for the first page? The api specification states `GET /address/:address/txs/chain[/:last_seen_txid]`. I guess you tested so it seems like it doesn't matter

theodorebugnet: Works just fine, and this way it doesn't need special-casing.

bitcoin/src/iter.rs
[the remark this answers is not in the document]
sander2: So is this handled somewhere currently? What would happen if the electrs server is down and one of the issues is in a pruned block? Do we restart or do we quietly continue?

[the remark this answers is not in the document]
theodorebugnet: Then we return and `Err()`, which does trigger a restart as afar as I can tell.

sander2 (approved):

PR #324, opened 2022-08-03
test: use instant-seal for tests

sander2:
This reduces the whole testing suite to only about 50 seconds on my machine. The changes in `rpc.rs` are hopefully temporary until [the underlying issue](https://github.com/paritytech/substrate/issues/10279) gets fixed

### DIFF
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -51,12 +51,13 @@ jobs:
           command: check
           args: --release --workspace --features ${{ matrix.metadata }}
       - name: test
+        if: matrix.metadata == 'parachain-metadata-kintsugi-testnet'
         uses: actions-rs/cargo@v1
         env:
           RUST_LOG: info,regalloc=warn
         with:
           command: test
-          args: --release --workspace --features parachain-metadata-kintsugi-testnet
+          args: --release --workspace --features ${{ matrix.metadata }}
       - name: upload artifacts - vault
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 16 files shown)


Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
```
sander2, line 2: This is required for correct feature selection. With resolver v1, if we had a dev-dependency on runtime with the `testing-utils`, it would also select the feature for non-dev builds. Version 2 correctly does the following:

> Features enabled on https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies will not be unified when those same dependencies are used as a normal dependency, unless those dev-dependencies are currently being built.

.github/workflows/cargo-test.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           RUST_LOG: info,regalloc=warn
         with:
           command: test
-          args: --release --workspace --features ${{ matrix.metadata }}
+          args: --release --workspace --features parachain-metadata-kintsugi-testnet
```
sander2, line 59: I wanted to run the tests only for `parachain-metadata-kintsugi-testnet` for now, but it looks like instead it causes this exact test to run once for each item in `matrix.metadata`. @ns212 do you know how to fix this?

sander2: @gregdhill Got this to work now. Let me know if you have any problems with running the tests on only one set of metadatas for now

gregdhill (approved):

PR #336, opened 2022-08-12
fix: restart child on crash, kill child on runner crash

savudani8:
- Restarts child process (vault) if crashed, checked in the main loop of `auto_update`, every `BLOCK_TIME`
- Kills child process before the runner shuts down because of an error. This is implemented both by implementing the `drop` trait for `Runner` and by calling `terminate_proc_and_wait()` if `auto_update()` ever returns

### DIFF
--- a/runner/src/runner.rs
+++ b/runner/src/runner.rs
@@ -179,6 +179,7 @@ impl Runner {
     }
 
     fn terminate_proc_and_wait(runner: &mut impl RunnerExt) -> Result<u32, Error> {
+        log::info!("Terminating child process...");
         let child_proc = runner.child_proc().as_mut().ok_or(Error::NoChildProcess)?;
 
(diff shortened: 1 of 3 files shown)


runner/src/runner.rs
```diff
@@ -310,6 +339,19 @@ impl Runner {
     }
 }
 
+impl Drop for Runner {
```
sander2, line 342: With the check that you have in `run`, is this still required? Having a long-running call (which the `_wait` implies it potentially is) inside a drop could potentially be quite confusing if you're not aware of it.

savudani8: I think many long-running functions implement `drop`, to clean up any open connections for instance. It's not required, but I like it as an added failsafe.

gregdhill (approved):

PR #340, opened 2022-08-25
feat(runner): upload artifact

savudani8:
Uploads `runner` build artifact so it can be released. Since it does not depend on any metadata feature it only runs once, on the `parachain-metadata-kintsugi-testnet` step (since that step is also the only one to run tests).


savudani8:
I see this warning in the pipeline but it seems like all other clients (except for the vault) also log it.
```shell
Warning: No files were found with the provided path: target/release/runner. No artifacts will be uploaded.
```

gregdhill (approved):

PR #353, opened 2022-09-06
feat(vault): handle termination signals

savudani8:
Handles termination signals, using a similar approach to https://github.com/interlay/interbtc-clients/pull/330.

This feature makes way for cleaning up the PID file on shutdown, in https://github.com/interlay/interbtc-clients/pull/347

I tried simplifying the scaffolding required for the mocks but still couldn't find a way around using trait methods only as an indirection to the actual implementation.

This feature is opt-in in the code (call `catch_signals()` instead of `start()`) and could be moved to the `service` crate instead.

---

Panics

    If the given signal is [forbidden][crate::FORBIDDEN].
    If the signal number is negative or larger than internal limit. The limit should be larger than any supported signal the OS supports.
    If the relevant Exfiltrator does not support this particular signal. The default [SignalOnly] one supports all signals. <- we're using the default

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13641,6 +13641,8 @@ dependencies = [
  "serde",
  "service",
  "sha2 0.8.2",
+ "signal-hook",
+ "signal-hook-tokio",
  "sp-arithmetic",
  "sp-core",
  "sp-keyring",
--- a/service/src/error.rs
+++ b/service/src/error.rs
@@ -2,6 +2,7 @@ use bitcoin::Error as BitcoinError;
 use hyper::{http::Error as HyperHttpError, Error as HyperError};
 use runtime::Error as RuntimeError;
 use serde_json::Error as SerdeJsonError;
+use std::io::Error as IoError;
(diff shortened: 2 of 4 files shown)


sander2:
> I tried simplifying the scaffolding required for the mocks but still couldn't find a way around using trait methods only as an indirection to the actual implementation.

One approach that wouldn't require trait methods is to to refactor like this:

```

savudani8:
> > I tried simplifying the scaffolding required for the mocks but still couldn't find a way around using trait methods only as an indirection to the actual implementation.
> 
> One approach that wouldn't require trait methods is to to refactor like this:
> 
> ```rust
> pub async fn catch_signals<F>(future: F)
> where
>     F: Future<Output = Result<(), Error>>,
> {
>     select (...)
> }
> ```
> 
> and then to call it as `catch_signals(start)`. In the tests you could just pass in a sleeping function

Much simpler, thanks!

gregdhill (approved):

sander2 (approved):

PR #364, opened 2022-09-13
feat(runner): CLI for client type, Readme update

savudani8:
Adds `client-type` CLI argument to the runner and updates the readme / comments.

I'm not sure whether using an `enum` that implements `FromStr` and `Display` is the cleanest approach, let me know what you think.


sander2 (approved):
lgtm

PR #365, opened 2022-09-13
feat(runner): don't re-download if checksums match

savudani8:
Uses checksum of the release to check if re-downloading can be avoided. Validation of the downloaded binary will be done in an upcoming PR.


sander2 (changes requested):
I think this still re-downloads the binary every time you start the runner, right? We should check if a binary with the correct hash already exists on disk

sander2 (approved):

PR #385, opened 2022-10-12
fix: don't remove pidfile if it wasn't created by current process

sander2:
I observed the following behavior:

- vault 1 starts, creates pidfile
- vault 2 starts, fails to create pidfile, exits, and removes pidfile 
- vault 2 starts again, now pidfile doesn't exist, causing it to start successfully

This PR addresses this by only removing the pidfile if the current instance attempted to write to the pidfile.


gregdhill (approved):
LGTM

savudani8 (approved):

PR #444, opened 2023-02-09
chore(faucet): update config files

daniel-savu:


gregdhill (approved):
Happy to have this change but we don't need to mirror the deployment files, they can just be examples of how to write the config.

PR #458, opened 2023-03-07
chore: lockfile check script

sander2:
A script that I wrote a while ago that checks for unpatched repos in the lockfile. Whenever we upgrade polkadot versions it's common that some crates need to be added to the patch list. This script prints such unpatches crates.


gregdhill:
Not needed https://github.com/interlay/interbtc-clients/pull/507

PR #530, opened 2023-10-04
refactor: add additional checks for dia fair price feed

gregdhill:
Check if all necessary variables exist (i.e. LockedToken, IssuedToken, Ratio, BaseAssetPrice)


sander2 (approved):
lgtm. What will happen if this fails though - will the other feeds still submit?

PR #536, opened 2023-11-22
fix: import keys into shared-v2 wallet

gregdhill:
Current logic flow now:

1. Fetch all currency specific Vaults
    1. Vault imports derivation key from master to shared and shared-v2
    2. Vault imports all keys from currency specific wallet to shared and shared-v2
    3. Sweep funds from currency specific wallet to shared
2. Add issue keys to shared wallet
    1. Rescan using electrs if using pruned node
    2. Rescan using bitcoind remaining heights
3. Add issue keys to shared-v2 wallet
    1. Rescan using electrs if using pruned node (only after last sweep - if it exists)
    2. Rescan using bitcoind remaining heights

### DIFF
--- a/vault/src/system.rs
+++ b/vault/src/system.rs
@@ -292,6 +292,7 @@ impl VaultIdManager {
 
         let btc_rpc_master = &self.btc_rpc_master_wallet;
         let btc_rpc_shared = self.btc_rpc_shared_wallet.clone();
+        let btc_rpc_shared_v2 = self.btc_rpc_shared_wallet_v2.clone();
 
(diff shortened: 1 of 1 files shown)


sander2 (changes requested):
I _think_ the change you made is correct, but (1) I think we need to clear rescan cache (by renaming, as we did before), otherwise it's not going to find the utxos of the imported addresses, and (2) it's been a while since I last looked closely into this code so it's hard to be 100% sure we didn't miss something..

PS you also need to cargo fmt

